PR #13, opened 2021-08-23
Attempt to clarify the language around proofs.

msporny:
This PR attempts to simplify the "in scope" work to include the definition of data models for proofs as well as the algorithms for the expression and verification of proofs that use existing cryptographic primitives.

### DIFF
--- a/index.html
+++ b/index.html
@@ -145,9 +145,9 @@ the Web.
 The <b>scope</b> of the Verifiable Credentials Working Group is:
         </p>
         <ul>
-          <li>Data models for verifiable credentials, verifiable presentations, and proofs</li>
+          <li>Data models for credentials, presentations, and proofs</li>
           <li>Registries for extending the data models</li>
-          <li>Algorithms for the expression and verification of proofs using existing cryptographic primitives</li>
+          <li>Algorithms for the expression and verification of proofs that use existing cryptographic primitives</li>
         </ul>
         <section id="section-out-of-scope">
           <h3 id="out-of-scope">Out of Scope</h3>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -206,7 +206,7 @@ <h3>
             </dd>
           </dl>
           <dl>
-            <dt id="linked-data-integrity-1" class="spec">Verifiable Credential Linked Data Integrity 1.0</dt>
+            <dt id="linked-data-integrity-1" class="spec">Linked Data Integrity 1.0</dt>
```
brentzundel: Adding "Verifiable Credential" to the name of this deliverable was _strongly_ recommended by Ivan to forestall some of the issues run into by the linked data signatures charter.

msporny: Ok, reverted.

I'm afraid that the name might cause people in the WG to hard code the solution to VCs, when the solution is intended to be generalized. I won't stand in the way here, but will raise the concern that people might take the wrong thing away from the title of the specification. I can live with it for now, and hope it won't blow up in our faces later. :)

iherman: To make my point clear: I do not want to see the same discussion coming up again from the LD community. If other communities pick up the VC solution for other problems than VC-s, good for them and for us. But if we just make hint that we have a solution for Linked Data at large we are back to square one. Other communities may vote with their feet, a.k.a. their code. It is not this wg's concern.

B.t.w., we should be consistent: in the other PR the deliverable's name in one of the liaison statement was also using the title of the spec without VC.

(we shoukd add the preview facility to this repo. I have difficulties to look PR-s on my iPad...)

iherman (approved):

brentzundel (approved):

Issue #21, opened 2021-11-23
Verifiable Credential Linked Data Integrity work should happen in a Linked Data working group


selfissued:
The new "Verifiable Credential Linked Data Integrity 1.0" deliverable (which was not present in the original charter), is sufficiently different from the core deliverable "Verifiable Credentials Data Model 2.0" that it should happen in its own working group.  Surely the Proposed Linked Data Signatures Working Group (https://w3c.github.io/lds-wg-charter/) would be a better place for this work - likely containing more of the relevant subject matter experts.

Please remove the second proposed deliverable from the draft charter.

selfissued:
And of course, the text "Algorithms for the expression and verification of proofs that use existing cryptographic primitives" should be moved from the Scope section to the Out of Scope section.

iherman:
That approach has been tried, and it does not seem to be workable. 

The LDS charter proposal[[1]](https://w3c.github.io/lds-wg-charter/) that you cite is, essentially, identical to the proposal that was put forward to the AC[[2]](https://lists.w3.org/Archives/Member/w3c-ac-members/2021AprJun/0002.html) and that was also submitted for review to the Linked Data/Semantic Web community[[3]](https://lists.w3.org/Archives/Public/semantic-web/2021Apr/0012.html). This was followed by a fairly lively email thread as a response to [[3]](https://lists.w3.org/Archives/Public/semantic-web/2021Apr/0012.html) (and also a more modest discussion on GitHub and the AC). After many weeks' of discussions, we had to conclude that there is no hope to find a consensus for [[1]](https://w3c.github.io/lds-wg-charter/), and submitting it to the AC would lead to formal objections. We acknowledged that to the Linked Data community in [[4]](https://lists.w3.org/Archives/Public/semantic-web/2021Oct/0020.html), as well as to the AC[[5]](https://lists.w3.org/Archives/Member/w3c-ac-forum/2021OctDec/0113.html).

[...]

PR #32, opened 2021-12-08
Clarify that VC data integrity is about more than just Linked Data.

msporny:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-wg-charter/pull/32.html" title="Last updated on Dec 8, 2021, 8:12 PM UTC (5ae139c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-wg-charter/32/bc165a4...5ae139c.html" title="Last updated on Dec 8, 2021, 8:12 PM UTC (5ae139c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -208,11 +208,10 @@ Exclusion period <b>began</b> 25 July 2019; Exclusion period <b>ended</b> 23 Sep
             </dd>
           </dl>
           <dl>
-            <dt id="linked-data-integrity-1" class="spec">Verifiable Credential Linked Data Integrity 1.0</dt>
+            <dt id="data-integrity-1" class="spec">Verifiable Credential Data Integrity 1.0</dt>
             <dd>
               <p>
-This specification defines how to express proofs of integrity for bounded Linked
-Data documents such as verifiable credentials.
+This specification defines how to express proofs of integrity, such as digital signatures or proofs of existence, for bounded documents such as verifiable credentials.
               </p>
 
               <p class="draft-status"><b>Draft state:</b> <a href="https://w3c-ccg.github.io/ld-proofs/">Community Group Report</a> </p>
@@ -244,8 +243,8 @@ Data documents such as verifiable credentials.
             </li>
             <li>Test suites for verifiable credential proof types:
             <ul>
-              <li>VC Linked Data Integrity - Ed25519</li>
-              <li>VC Linked Data Integrity - BBS+</li>
+              <li>VC Data Integrity - Ed25519</li>
+              <li>VC Data Integrity - BBS+</li>
               <li>VC JSON Web Tokens</li>
             </ul>
             </li>
@@ -314,7 +313,7 @@ testing for the specifications.
             <dl>
                 <dt>RDF Canonicalization and Hashing Working Group <span class="todo">Note: This WG is currently being chartered</span></dt>
                 <dd>
-                    To synchronize on canonicalization output expression mechanisms that might be used by the VC Linked Data Integrity specification.
+                    To synchronize on canonicalization output expression mechanisms that might be used by the VC Data Integrity specification.
                 </dd>
 
                 <dt><a href="https://www.w3.org/2019/did-wg/">Decentralized Identifier Working Group</a></dt>
@@ -344,11 +343,11 @@ testing for the specifications.
                 </dd>
                 <dt><a href="https://datatracker.ietf.org/rg/cfrg/about/">Internet Engineering Task Force Crypto Forum Research Group </a> </dt>
                 <dd>
-                    To perform broad horizontal reviews on the output of the Working Group and to ensure that new pairing-based and post-quantum cryptographic algorithms and parameters can be integrated into the Linked Data Integrity ecosystem.
+                    To perform broad horizontal reviews on the output of the Working Group and to ensure that new pairing-based and post-quantum cryptographic algorithms and parameters can be integrated into the Data Integrity ecosystem.
                 </dd>
                 <dt><a href="https://www.nist.gov/"> National Institute of Standards and Technology, U.S. Department of Commerce </a></dt>
                 <dd>
-                    To coordinate in ensuring that new pairing-based and post-quantum cryptographic algorithms and parameters can be integrated into the Linked Data Integrity ecosystem.
+                    To coordinate in ensuring that new pairing-based and post-quantum cryptographic algorithms and parameters can be integrated into the Data Integrity ecosystem.
                 </dd>
                 <dt><a href="https://www.hyperledger.org/use/aries"> Hyperledger Aries</a></dt>
                 <dd>


brentzundel (approved):

kdenhartog (approved):

iherman:
These are questions that come from a layperson (I am far from being familiar with all the minute details of the VC spec, and I may be wrong). Looking at the current standard, what I see is:

- [§6.3.1 JSON Web Token](https://www.w3.org/TR/2021/REC-vc-data-model-20211109/#json-web-token) is a normative section that seems to define the proof format with JWT. It seems fairly complete, so the question that might be raised is: what is the charter proposing in this area? Why is it necessary to reopen this section? 
- [§6.3.2 Linked Data Proofs](https://www.w3.org/TR/2021/REC-vc-data-model-20211109/#linked-data-proofs) is indeed, not formally normative (it relies on drafts, ie, non-normative specs). (As an aside, how come the section is not marked as non-normative?) The charter is clearly meant to upgrade that sections (or the references therein).

So what are we talking about as _non-LD_ work? Are we talking about a non-JWT and non-LD form of proof? If so, what?

msporny:
> * [§6.3.1 JSON Web Token](https://www.w3.org/TR/2021/REC-vc-data-model-20211109/#json-web-token) is a normative section that seems to define the proof format with JWT. It seems fairly complete, so the question that might be raised is: what is the charter proposing in this area? Why is it necessary to reopen this section?

The section contains the following problems:

* Numerous inconsistencies (as implementers have discovered)
* It doesn't belong in the VC Data Model Spec, but an external spec that talks about securing a VC using a JWT.

> * [§6.3.2 Linked Data Proofs](https://www.w3.org/TR/2021/REC-vc-data-model-20211109/#linked-data-proofs) is indeed, not formally normative (it relies on drafts, ie, non-normative specs). (As an aside, how come the section is not marked as non-normative?) The charter is clearly meant to upgrade that sections (or the references therein).
> 
> So what are we talking about as _non-LD_ work? Are we talking about a non-JWT and non-LD form of proof? If so, what?

It is possible to create a proof that is not RDF-based and not JWT-based. For example, if you use the [ld-proofs](https://w3c-ccg.github.io/ld-proofs/) and canonicalize using [RFC8785](https://datatracker.ietf.org/doc/html/rfc8785), you can create a clear-text signature that uses the proof format described in [ld-proofs](https://w3c-ccg.github.io/ld-proofs/). There have been a few proposals in this vein in the CCG. Unfortunately, when people see the "LD" part of the specification name, they jump to the wrong conclusion that you MUST use RDF... which is just completely false.

JSON-LD used the term "Linked Data" in the non-RDF sense (before the RDF WG got a hold of the spec and made the mapping to RDF normative). "Linked Data" in "Linked Data Proofs" was meant to convey "a graph-based data structure that uses hyperlinks"... but, unfortunately, people holding extreme positions on both sides of the spectrum are choosing to insist that  "LD" means more than just "a graph-based data structure that uses hyperlinks".

iherman:
@msporny I am fine with all this, but it may be worth hinting at some of this in the charter. Adding something that makes it clear that the new recommendation _may_ override/replace/complement some sections in the core data model spec, that we are defining a deliverable that may affect the core and not only a more-or-less independent specification. Otherwise, we will face questions like the ones I was asking...

msporny:
> Adding something that makes it clear that the new recommendation _may_ override/replace/complement some sections in the core data model spec, that we are defining a deliverable that may affect the core and not only a more-or-less independent specification.

Good suggestion, I'll add that text to this PR.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2021-12-15-vcwg#section1-2) on 2021-12-15

- no resolutions were taken

#### 1.2. Verifiable Credential Linked Data Integrity work should happen in a Linked Data working group (issue vc-wg-charter#21)

_See github issue [vc-wg-charter#21](https://github.com/w3c/vc-wg-charter/issues/21)._




**Brent Zundel:** manu I believe you have a PR open for this. Can you tell us about this?.  

_See github pull request [vc-wg-charter#32](https://github.com/w3c/vc-wg-charter/pull/32)._




**Manu Sporny:** yeah there's a PR open for this.  
… what it's trying to do is make it clear that this next spec will be talking about data integrity for verifiable credentials in a more wholistic and normative way.  
… this means JWTs/LD-Proof and any other mechanism would be in scope as well.  
… including things like JCS and other things like that.  
… we can't invent new crypto here but can focus on how things get packaged up.  
… by calling this LD Integrity it's not accurately representing this.  
… the specific reasons we're doing this is because we want to be clear that we're not just talking about Linked data integrity and making this about VC data integrity.  

**Brent Zundel:** Assuming the JWP work gets further along would you expect we could use that in here?.  

**Dave Longley:** My one concern with this PR is that we don't want to convey that we're only doing JWTs and want to make it clear the Linked Data proofs are in scope.  

**Manu Sporny:** One thing we do to counteract that is to link specifically to the linked data proof spec.  
… I can check in CCG about renaming but this should make it clear that we want it in scope.  
… I'll add some clarifying language to explicitly list the things in scope.  

**Kyle Den Hartog:** I was thinking about... the proof section becomes obvious, I'm not sure to what degree we wanted to scope particular suites... with JWTs you don't hit that as much, how do we want to handle scoping of that?.  

**Manu Sporny:** I believe the way it's scoped now is that we say all of the suites are non-normative.  
… one second that might not be true.  
… we don't say anything about the suites themselves.  
… I think the intent was to note define specific suites.  
… we'd publish notes and test suites for it but not define them normatively.  

> *Dave Longley:* -1 we shouldn't constrain ourselves to not being able to define any suites.

**Manu Sporny:** but that might be the wrong way to go about it.  

**Kyle Den Hartog:** I'm concerned that suites are being developed now in a way that makes it problematic..  
… Like there are 2020, 2021, 2022 suites being developed, taking longer than a year to stabilize -- we need some sort of model to encompass the active work happening right now..  

**Dave Longley:** I don't think we should constrain ourselves to not being able to address these suites.  
… we may be able to develop a versioning scheme that's a bit of a middle ground where we create a few with common vocabs to describe these.  
… and that might avoid some of the issues that Kyle is bringing up.  
… we might be able to handle this, but my main point is that I don't think we should constrain ourselves to solve that.  

**Manu Sporny:** There is a VC extension registry.  
… I think we're probably going to end up making that in scope.  
… I'd imagine doing all that to document that things are in process.  
… to dlongley point I think the big question is how normative do we want to get about the suites.  
… I think the idea was that we'd define a bunch of the suites as notes.  
… part of the DID-Core objections are that you should have just taking things to CR and not gone further.  
… so if we take that into consideration for this chartering we may follow a similar model.  
… where we only take a suite to CR without saying we're going to work on all of these suites normatively.  
… and stay focused on a few that we take to CR and if we have good enough interop we can ask W3C to take it to REC later.  
… this would allow us to make it clear that we're not overscoping our work.  
… so we need to be careful about what we say we're going to do.  
… this is to say we'll change the description to say we'll describe Edwards curve, BBS+, and JWTs but not take them fully to REC.  

**Brent Zundel:** How common is it for people to say we're only going to CR only?.  

**Manu Sporny:** Not common it's only for CSS typically and other perma working groups.  

**Brent Zundel:** with just my hat on I'm not sure going to only CR makes sense.  

**Dave Longley:** +1 I think it's best to say we're going to handle only a few suites.  

> *Kyle Den Hartog:* Saying going only to PR seems like people could object to that.

**Manu Sporny:** I think we could do that for Edwards curve but could be trickier for BBS+.  
… we'd need CFRG approval before we could reference them normatively and succeed in REC stage.  

**Dave Longley:** if we go with something a bit more general we could describe how you use a suite without having to specifically normatively define BBS+.  

**Brent Zundel:** There's active work at DIF and they expect to have something to submit to the CFRG first quarter next year.  

**Manu Sporny:** Is there something we can point out?.  
… even if it's an ID that intends to go to CFRG.  

> *Brent Zundel:* See [current draft of BBS+](https://identity.foundation/bbs-signature/).

**Manu Sporny:** I think the response to the JWTs handling all of this is that you can't do multiple signatures together with JWTs today.  
… the other issue is that we have a parallel RDF canonicalization WG.  
… and that group is expecting the ability to not have to use only JWTs only.  
… and with people already using this I think we have sufficient counter arguments for these things being brought up.  
… and we've documented all that stuff.  
… I think we're ok, but we'll see how it goes.  

**Kyle Den Hartog:** I think that's a sufficient argument.  

**Manu Sporny:** I guess the question is what do we want to do for this..  
… Do we want to update the description for this PR and describe focusing on Edwards and BBS+.  
… and because this is a normative item we can then decide as a WG to decide how we want to split up that work.  
… we'd have one description to define how to express VC integrity with a specific focus on Ed25519, BBS+, and JWTs.  

> *David Chadwick:* +1.

> *Dave Longley:* +1 to what manu just said and to pointing to this discussion saying we meant to define LD suites.

**Manu Sporny:** and if there's ever a question about whether or not we meant to include Linked Data we can point back to these minutes to show we were very clear.  

**Brent Zundel:** So what needs to change for that to happen?.  

**Manu Sporny:** I need to change this PR to include it and I'll link to draft state input documents to cover this.  

</details>

iherman (approved):

PR #42, opened 2022-01-20
Unambiguously state that fixing problems is in scope

### LINKED ISSUE #22: Fixing problems identified through using the 1.0 and 1.1 specs must be in scope
Several problems have been identified when people tried to use the 1.0 specification.  Some of these are ambiguities, resulting in different interpretations by different implementers.  Some of them are conflicts between normative text.  All potentially reduce interoperability among implementations.  Fixing these problems needs to be explicitly within scope, even in cases where the fix involves changing normative requirements relative to the previous specifications.

Please make fixing problems present in the earlier specifications explicitly within scope.

selfissued:
Fixes #22

### DIFF
--- a/index.html
+++ b/index.html
@@ -146,12 +146,15 @@ the Web.
 The <b>scope</b> of the Verifiable Credentials Working Group is:
         </p>
         <ul>
-          <li>Addressing errata found in previous versions of the Verifiable Credentials Data Model</li>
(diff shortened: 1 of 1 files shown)


kdenhartog (approved):

iherman (approved):

iherman:
@Sakurann, I cannot yet call on you for review (per github's interface) but I would prefer you did it nevertheless...

brentzundel (approved):

Sakurann (approved):

Issue #38, opened 2022-01-11
Multilingual support for selecting a image file


shigeya:
I found use cases that selectively refer to image files that include strings in different languages as part of the images.

A change to this text in the charter is necessary:

> Extensions for binding multilingual strings to GNU gettext-like localization files

I couldn't immediately come up with a PR.
If there is any suggestion for change, please provide it.

msporny:
This was addressed by https://github.com/w3c/vc-wg-charter/pull/33.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-01-12-vcwg#section2-2) on 2022-01-12

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 2.2. Multilingual support for selecting a image file (issue vc-wg-charter#38)

_See github issue [vc-wg-charter#38](https://github.com/w3c/vc-wg-charter/issues/38)._




**Shigeya Suzuki:** Sometimes we need to select from multiple image files to handle multi lingual text. Images may contain text inside images..  
… This is different from how multilingual text is handled.  
… I was thinking about opening a PR, but I could not find a good way to disclose that in a simple text.  

**Brent Zundel:** Manu posted a comment.  

**Manu Sporny:** But I was wrong.  

**Brent Zundel:** Any suggestions for shigeya how to address this?.  

**Manu Sporny:** We should reach out to Internationalization folks. We might want to ask them. Nothing is coming to mind, I understand the problem and desired to address it, but I don't know...  
… I could think of one way in JSON-LD how to do it, but not in this text approach..  
… Maybe Internationalization folks have an idea how this can be done.  

**Shigeya Suzuki:** I will talk to my friend to find out who is best to talk to.  

**Brent Zundel:** As you discover things, please add comments to the issue.  

**David Chadwick:** Work Human Colossus is doing on OCA could be a solution for this. They have tables with different tables to represent things. Table could also contain multiple images..  
… I think markus is familiar with this.  

**Markus Sabadello:** Little bit, but not enough to solve this.  

**David Chadwick:** I'll give you an email address of Human Colossus lead, you can chat to him directly. There are various presentations on Google, Youtube...  
… On ESSIF-Lab they did a demo a few weeks ago.  

> *David Chadwick:* paul.knowles@humancolossus.org for details of Overlays Capture Architecture (OCA) that should provide a solution to multi-lingual images.

[...]

Issue #19, opened 2021-11-10
Standardization of Multilingual Support


shigeya:
As I mentioned during the TPAC 2022 VC meeting two weeks ago, we need multilingual support, especially for Asian countries.

While some examples (even the first example) in the Verifiable Credential  1.1 spec show how you can encode multinational strings into a property value, there is a lack of discussion about this scheme.

Since these strings need to be content-integrity protected, externalizing translation mapping is not an option; multilingual support must be part of the Verifiable Credentials Data Model.

In Japan, we already observe difficulty in vaccination certificates since we need to use English for overseas purposes and Japanese for domestic purposes. Both the data model and the software support are in need. Thus, standardizing support in the Verifiable Credentials Data Model is necessary.

We understand this is a challenging topic, especially when looking at the data representation syntax point of view. But for the more general, international adoption, we need multilingual support without affecting single-lingual syntax.

I think including this work as part of Verifiable Credential 2.0 work is essential.

shigeya:
I forgot to mention that I (of course) want to contribute. I'm thinking about how this work fits into the charter and looking forward to having discussions/comments.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2021-12-01-vcwg#section1-1) on 2021-12-01

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 1.1. Standardization of Multilingual Support (issue vc-wg-charter#19)

_See github issue [vc-wg-charter#19](https://github.com/w3c/vc-wg-charter/issues/19)._




**Brent Zundel:** Seems like there is support for this, how do we make sure this happens?.  

**Shigeya Suzuki:** I was not sure how to proceed w/ PR -- how do we fit this into the charter?.  
… I'll propose something in a PR..  

</details>

kdenhartog:
I'm curious if the need for this extends beyond the support offered by JSON-LD today or if this is something that you're looking to get standardized among all of the different representations? Also worth asking, is the usage of JSON-LD's feature for this fall under your classification of "externalizing translation mapping"?

shigeya:
@kdenhartog thanks for the comment.  Sorry for the late response.

Feature-wise, for the use-cases I saw, the language-indexing feature mentioned in [Language indexing](https://json-ld.org/spec/latest/json-ld/#language-indexing) of JSON-LD 1.1 is sufficient.

[...]

PR #44, opened 2022-01-24
Extending support for resources in general

shigeya:
Relates to Issue #38 

- Supports non-string resources such as images
- References to gettext removed to avoid confusion

***
<a href="https://pr-preview.s3.amazonaws.com/shigeya/vc-wg-charter/pull/44.html" title="Last updated on Feb 9, 2022, 9:44 AM UTC (62f9866)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-wg-charter/44/9a2a8d1...shigeya:62f9866.html" title="Last updated on Feb 9, 2022, 9:44 AM UTC (62f9866)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -270,7 +270,7 @@ produced.
               <li>VC JSON Web Tokens</li>
             </ul>
             </li>
-            <li>Extensions for binding multilingual strings to <a href="https://www.gnu.org/software/gettext/" title="GNU gettext">GNU gettext</a>-like localization files</li>
+            <li>Extensions for binding multilingual resources for localized user interfaces.</li>
           </ul>
           <p>
 The Working Group may also update Notes published under previous charters.


iherman (approved):

iherman:
@sakurann, can you review this as well?

Sakurann (changes requested):
It was difficult for me to understand what the work item is about from the first read of this sentence. After reviewing the related issues (including #19), I understood that the intent is to support multilingual representations. Let me know if there is a specific reason to specify string vs resource, or the mechanism of binding (what to what?), which we can discuss once the charter is approved. Otherwise, I would suggest alternative language, something like:
"Support for multilingual representations". (Or  "Support for multilingual claim representations" if we want to be specific.)

shigeya:
@Sakurann The mapping can be handled by either library which handles the data model or the user interfaces. It depends on how the software selects the languages. It is possible to show multiple languages on a screen simultaneously or show the text in the predetermined languages (by system settings).

We need to distinguish resource references from text strings. If a property expects strings, the value should be strings. But if a property expects a URL, then the value type is a string, but it might need special handlings.

Anyway, I'm OK with the language "Support for multilingual representations," if it is clear enough for other members.

Sakurann:
I understand there has been discussions and investigations around supporting multilingual representations for both resources and text strings, and that this is important for this use-case. How about the following text?
> Support for multilingual representations of resources not limited to the text strings

shigeya:
> Support for multilingual representations of resources not limited to the text strings

Clearer. Thanks. I will update the PR

shigeya:
> Support for multilingual representations of resources not limited to the text strings

I received feedback that this language gives the impression that the resource is possibly in the VC itself.
(Do we want to embed images into VC? It looks odd, but it seems there are valid use cases)

In that light, We (the guy who provided the feedback and me) still think "binding" is a suitable word. The resources will be included as part of the VC or linked to an external resource.

I am still thinking. Feedbacks welcome.

Sakurann:
So you want to provide multilingual support for the resources that are included in a credential by reference and not by value? 
Suggestion below:
> Support for multilingual representations of resources not limited to the text strings that can be included by reference or by value

shigeya:
> So you want to provide multilingual support for the resources that are included in a credential by reference and not by value? Suggestion below:
> 
> > Support for multilingual representations of resources not limited to the text strings that can be included by reference or by value

I mean both on reference or value for each of the properties.

OR13 (approved):

David-Chadwick:
grammatical change request please
to the text strings
change to
to text strings

shigeya:
> grammatical change request please 
> to the text strings 
> change to
> to text strings

Thank you. will include the change.

"Support for multilingual representations of resources not limited to text strings that can be included by reference or by value"

shigeya:
PR Updated. 

Now seeing "Error: 429 Too Many Requests" by PR Review. Is this okay... ?

(I found I'm working on the main branch in my repository:< this should not be the cause of the above issue...)

index.html
```diff
@@ -270,7 +270,7 @@ <h3>
               <li>VC JSON Web Tokens</li>
             </ul>
             </li>
-            <li>Extensions for binding multilingual strings to <a href="https://www.gnu.org/software/gettext/" title="GNU gettext">GNU gettext</a>-like localization files</li>
+            <li>Extensions for binding multilingual resources for localized user interfaces.</li>
```
Sakurann, line 273: The change I was proposing was for this string...

index.html
```diff
@@ -150,7 +150,7 @@ <h2>Scope</h2>
           <li>Data models for credentials, presentations, and proofs</li>
           <li>Registries for the data models</li>
           <li>Algorithms for the expression and verification of proofs that use existing cryptographic primitives</li>
-          <li>Refining multilingual support in data models</li>
+          <li>Support for multilingual representations of resources not limited to text strings that can be included by reference or by value"</li>
```
Sakurann: There is `"` left at the end btw.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-01-26-vcwg#section7-1) on 2022-01-26

- no resolutions were taken

#### 7.1. Multilingual support for selecting a image file (issue vc-wg-charter#38)

_See github issue [vc-wg-charter#38](https://github.com/w3c/vc-wg-charter/issues/38)._




**Brent Zundel:** "Multilingual support for selecting a image file" Raised by Shigeya..  
… There is conversation in the PR.  

_See github pull request [vc-wg-charter#44](https://github.com/w3c/vc-wg-charter/pull/44)._




**Shigeya Suzuki:** The original markdown... not only the text but also the image file....  
… This PR and the last one, for multilingual text support, replacing strings....  
… to support both localized string and localized images..  
… I had some conversation... I hope she understands my intention....  
… Committed by some folks at Keio.  
… I'm still working with Kristina on some of these changes.  
… If she is satisfied with that, I'll create a PR.  

**Kristina Yasuda:** Are you suggesting to keep the original PR, or accepting the change request?.  

**Shigeya Suzuki:** I'm a bit confused about [your comment](https://github.com/w3c/vc-wg-charter/pull/44#issuecomment-1022484682) 2 hours ago.  
… The support for multilingual is not limited to ...?.  

> *Kristina Yasuda:* "Support for multilingual representations of resources not limited to the text strings that can be included by reference or by value".

**Brent Zundel:** Thank you both.  

</details>

index.html
shigeya: Updated. thanks...

index.html
shigeya: Ah, right... my bad.

shigeya: This part of the text (in 2.2 Other Deliverables) lists a non-normative document on external mapping objects which we develop. So I think this should be ok.

I thought your request was for Section "1. Scope"

index.html
```diff
@@ -150,7 +150,7 @@ <h2>Scope</h2>
           <li>Data models for credentials, presentations, and proofs</li>
           <li>Registries for the data models</li>
           <li>Algorithms for the expression and verification of proofs that use existing cryptographic primitives</li>
-          <li>Refining multilingual support in data models</li>
+          <li>Support for multilingual representations of resources not limited to text strings that can be included by reference or by value</li>
```
msporny: I find this text, while specific, still vague and confusing. Do we mean "RDF resources" or "Web resources" or "anything at the end of a URL"? Are we expressing the references as URLs?

Feels like less is more here and what we had before seemed fine. It feels like the second change request here is good -- it describes a goal "Extensions for binding multilingual resources for localized user interfaces." -- can we try something like that here?

msporny (changes requested):

index.html
shigeya: Thanks for the comment. I do not have enough brain cycle to think about this week, and I can't attend the VCWG teleconference. I welcome suggestions. If there are not, I will figure them out early next week.

brentzundel: I think that the text we had was sufficient, and covers what is described here.
I would recommend keeping the original text.

index.html
brentzundel: @Sakurann could you suggest text which you would find acceptable here?

index.html
shigeya: > I think that the text we had was sufficient, and covers what is described here. I would recommend keeping the original text.

Okay, let me revert the change on this.

brentzundel (approved):

brentzundel (approved):

Sakurann (approved):

Issue #48, opened 2022-01-25
Standardizing PGP for Linked Data Integrity


OR13:
https://w3id.org/security/suites/pgp-2021
https://w3id.org/security/suites/pgp-2021/v1

PGP / GPG are commonly used in software supply chain uses cases, and even in hardware thanks to [Yubikey / Open PGP](https://support.yubico.com/hc/en-us/articles/360013790259-Using-Your-YubiKey-with-OpenPGP).

GitHub integrations with [GPG](https://docs.github.com/en/authentication/managing-commit-signature-verification/adding-a-new-gpg-key-to-your-github-account)

[Signing Docker images with GPG](https://www.unboundsecurity.com/docs/UKC/UKC_Code_Signing_IG/HTML/Content/Products/UKC-EKM/UKC_Code_Signing_IG/GPG/GPGdocker.htm)

While PGP won't play nice with VC-JWT, it can play nice with LD Proofs, as shown the specs at the top of this issue.

You can even use GPG Agent to create verifiable credentials that are bound to hardware isolated keys (such as with yubikey).

Is there interest in standardizing support for PGP / GPG based Linked Data Integrity suites?

mprorock:
Absolutely.  This could be highly valuable to us

OR13:
Here is an example CLI that works with these suites, https://github.com/OR13/lds-pgp2021/blob/main/bin/cli.js

it uses GPGAgent and Yubikey, and allows you to make a verifiable credential that is signed by a hardware isolated key on yubikey but verifiable to any DID that supports PGP. (or in the context of VCs any controller).

OR13:
Open PGP supports most of the cryptography that is popular these days.

Checkout https://openpgpjs.org/

> This project is maintained by ProtonMail.

https://github.com/openpgpjs/openpgpjs

See also: https://github.com/openpgpjs/openpgpjs#security-audit

wyc:
+1, we are supporting PGP keys as well:
https://github.com/spruceid/ssi/pull/373

brentzundel:
I believe that merging PR #77 will fix this issue

brentzundel:
I believe this issue has been addressed. Marking as pending close.
I will close this issue in 4 days if there are no objections.

brentzundel:
no response since marked `pending close`, closing

PR #50, opened 2022-01-26
Add a possible reference for pgp as a cryptosuite

mprorock:
in reference to #48

***
<a href="https://pr-preview.s3.amazonaws.com/mesur-io/vc-wg-charter/pull/50.html" title="Last updated on Jan 26, 2022, 8:47 PM UTC (8702138)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-wg-charter/50/9a2a8d1...mesur-io:8702138.html" title="Last updated on Jan 26, 2022, 8:47 PM UTC (8702138)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -221,7 +221,11 @@ credentials. Concrete serializations will be provided based on <a
 href="https://w3c.github.io/vc-data-model/#json-web-token">VC-JWT</a>, the <a
 href="https://w3c-ccg.github.io/lds-ed25519-2020/">Ed25519 Cryptosuite</a>, and
 the <a href="https://w3c-ccg.github.io/lds-ecdsa-secp256k1-2019/">Secp256k1
-Cryptosuite</a>. Concrete serializations based on the <a
+Cryptosuite</a>. 
+Concrete serializations based on the <a
+href="https://or13.github.io/lds-pgp2021/">PGP Cryptosuite</a> might be provided,
+if a complete specification is published in time.
+Concrete serializations based on the <a
 href="https://w3c-ccg.github.io/ldp-bbs2020/">BBS+ Cryptosuite</a> might be provided, if an <a
 href="https://identity.foundation/bbs-signature/">IETF BBS+ RFC</a> is published
 in time. Concrete serializations for VC-JWP might be provided, if an <a
@@ -238,6 +242,7 @@ produced.
                 <a href="https://w3c-ccg.github.io/lds-ed25519-2020/">Ed25519 Cryptosuite</a>,
                 <a href="https://w3c-ccg.github.io/lds-ecdsa-secp256k1-2019/">Secp256k1 Cryptosuite</a>,
                 <a href="https://w3c-ccg.github.io/ldp-bbs2020/">BBS+ Cryptosuite</a>
+                <a href="https://or13.github.io/lds-pgp2021/">PGP Cryptosuite</a>
               <p class="milestone"><b>Expected completion:</b> Q4 2023</p>
             </dd>
           </dl>


msporny (changes requested):
-1 to this PR, for the following reasons:

1. This links to a specification that is not covered under any W3C IPR regime.
2. The work on PGP cryptosuites has not received much, if any, interest outside of the maintainer.
3. There are no vendors that are committing to supporting this cryptosuite in production at this time. 
4. PGP is dead, with a short list of citations [1](https://www.vice.com/en/article/vvbw9a/even-the-inventor-of-pgp-doesnt-use-pgp), [2](https://www.cryptologie.net/article/487/a-history-of-end-to-end-encryption-and-the-death-of-pgp/), [3](https://www.wired.co.uk/article/efail-pgp-vulnerability-outlook-thunderbird-smime), [4](https://news.ycombinator.com/item?id=28376644), ... with many more if you just google for opinions on PGP.

mprorock:
> 3. There are no vendors that are committing to supporting this cryptosuite in production at this time.

see issue: #48 - I see two vendors: Spruce (@wyc), and Transmute (@OR13) expressing support, with a third: mesur.io (myself) stating an interest in this

msporny:
> I see two vendors: Spruce (@wyc), and Transmute (@OR13) expressing support

I see two vendors supporting an experiment, not stating that they're going to support this in production. That said, if there are enough vendors that are willing to support this in production, and put in the time to do the spec and test suite, I'm certainly not going to block it (even though I don't think it's the best use of time in the VCWG given all of our other priorities).

OR13:
We should merge this PR ASAP so we can take up cleaning up the entire section, see: https://github.com/w3c/vc-wg-charter/issues/54

OR13:
@msporny this is a charter PR... I think its best to retain the ability to speak to popular crypto suites such as JWS / CWS / PGP in the charter... otherwise expect formal objections at the end... we prefer to see them at the beginning of the work.

I also think we need to do a better job of standardizing how work outside the W3C can leverage these standards, similar to how we defined did spec registries for extensions.

In order to do that job well, we need to be able to discuss items like this in the WG.

msporny:
> In order to do that job well, we need to be able to discuss items like this in the WG.

I would be a +1 for a PGP Cryptosuite to be put into the non-normative section of the charter as well as defining a registry for all cryptosuites (which I think is already in scope). I do agree that we should enable it to exist. I am concerned about spending time to normatively define it while trying to do everything else in the WG.

OR13:
@msporny are you asserting that there will be a normative defintion for some registered crypto suites, but that a registry should exist to support suites that are "not normatively defined"?

msporny:
> @msporny are you asserting that there will be a normative defintion for some registered crypto suites, but that a registry should exist to support suites that are "not normatively defined"?

I'm suggesting that there should be a registry for cryptosuites. That registry might break those registrations into "SDO Standards", "Experimental Implementations", and "Deprecated" (or whatever classifications folks feel are acceptable)... but I do think that we should have a registry that points to everything we know of in the ecosystem. We're going to want to do something different than what we did w/ the DID Spec Registries... and again, this is just a suggestion that needs to be debated in the WG. My hope is that if PGP doesn't make it into the WG as a normative thing, that it would be worked on as a non-normative thing, and referred to in the registry so that people that want to use it can use it. Given that there are at least two implementers for PGP already, with a third one in the wings, I expect that there is a case for making it a normative work item.

OR13:
> My hope is that if PGP doesn't make it into the WG as a normative thing, that it would be worked on as a non-normative thing, and referred to in the registry so that people that want to use it can use it.

Right, I am hearing there will be a registry and it will point to normative and non normative things... that is exactly like did spec registries (it points to did core properties, and it defines non normative extensions).

I am not actually suggesting that a PGP be a normative work item, I am just trying to make sure we have enough useful examples to navigate the "registries conversations"... and to gather understanding regarding the view on "normative" vs "non normative" suites in the WG.

msporny:
> I am not actually suggesting that a PGP be a normative work item

Ah! So, the section that this PR places the PGP work item in is for normative work items for the VC WG. This PR is explicitly stating that a PGP cryptosuite should be a normative work item.

OR13:
> Ah! So, the section that this PR places the PGP work item in is for normative work items for the VC WG. This PR is explicitly stating that a PGP cryptosuite should be a normative work item.

The entire section needs to be rewritten, as I noted here:

https://github.com/w3c/vc-wg-charter/issues/54

All the items are tangled together, labeled incorrectly, not enough information to evaluate...

The sooner we accept these PRs to add list items, the sooner we can clean up the list.

OR13 (approved):
I think we should allow the WG to explore PGP since it is the dominate technique to securing software supply chain.

msporny:
> I think we should allow the WG to explore PGP since it is the dominate technique to securing software supply chain.

Of course, there is no objection (that I can see) to that happening. The only point of contention is specifying a PGP cryptosuite normatively vs. non-normatively. To be clear, I would be a +1 to defining a PGP cryptosuite non-normatively in preparation for doing it normatively in a VC 3.0 WG OR, having some other group work on it normatively and referring to that specification in the cryptosuite registry. I'm concerned about adding more normative work items other than things that are in the critical path for multiple vendors that are participating in the WG.

I hope this helps clarify things. If not, I can raise another PR for PGP that places it as a non-normative cryptosuite deliverable for the WG to define.

kdenhartog:
It's my take that PGP support would be a good example of an crypto suite that would work well for an extension, but I also think that it's going to introduce some odd couplings of the way in which the proof data model is defined, how an implementation is needed, how public keys are resolved, and how keys are formatted. In my opinion it's the VCWG job to define only the proof data model which is why if all of these aspects end up necessary to sufficiently define it then I believe it should be defined as an extension rather than normatively defined by the WG.

OR13:
@kdenhartog yeah, I agree, I think it will be helpful to define at least 1 extension. However it's not clear to me how best to account for the extension registry and examples in the current charter.


I would be in favor of moving this to the non normative section.

kdenhartog:
Personally I don't think things in the extension registry need to be listed in the charter unless it's going to be REC track. I think if the WG decides to make this a note or register the CCG report then either approach wouldn't necessitate a charter since it wouldn't affect exit criteria for the WG. If we take this approach as well it should resolve things like #72 without further considerations being needed.

iherman:
> Personally I don't think things in the extension registry need to be listed in the charter unless it's going to be REC track. 

With the new W3C process, a registry can be a simple one or a [W3C Registry track document](https://www.w3.org/2021/Process-20211102/#registries), which does involve its own process. I believe that, while you may be right for the former, we should add an item in the charter if our goal is the latter.

kdenhartog:
I see what you're saying for the note and agree that this sounds correct. However, since a note by the working group isn't protected by the patent policy (see the Green highlighted point in [section 6.4.1 of the process document](https://www.w3.org/2021/Process-20211102/#Note)) I actually think that a note is less useful than a CCG report since that has a CLA to protect it. Since this is the charter I don't want to debate this too much further until we can discuss in the WG, but here's what I tentatively would have in mind.

1. Each registry table requires a good definition for a registry entry
2. Registry entries that are "required" status MUST be normatively defined as REC track documents (think default options)
3. Registry entries that are "recommended" or "optional" status MUST be published in a royalty free manor (e.g. REC track docs or CCG reports - wording could probably be improved)
4. Registry entries that are "provisional" status MUST be converted to "recommended" or "optional" status within 1 year or request an extension from the Registry custodian

With these rules in mind, I'd think something like PGP cryptosuite would best fit under a CCG published report with "optional" status since it gives a greater protection with regards to patent policies, allows for it to be moved into TR space after stabilization in a later WG as a rec track document, while still allowing for much of the work to be stabilized under normal processes. The tough part I see here is I don't actually think we'd be able to reach consensus on a "required" status crypto suite since we'd likely run back into the JWT vs JSON-LD debate and I could see that being used as evidence against the WG to object. For certain registry tables I think that'd be fine (e.g. an "evidence" registry table), but for cryptosuite registry tables I doubt that would fly during a review. 

In any case, I think having a better understanding of the way the registry process will work from the outset will be crucial to informing what should be put normatively in scope and the rest I actually think is best left to the CCG for patent reasons and this is a perfect example of one that I think would fall firmly in the stay in CCG, finalize the CCG report there and register as optional status category.

OR13:
> The tough part I see here is I don't actually think we'd be able to reach consensus on a "required" status crypto suite since we'd likely run back into the JWT vs JSON-LD debate and I could see that being used as evidence against the WG to object.

I think we should not run from objections we should run towards them, and address them.

I am sure that folks will try to make the case for a few required suites, and if we can't agree on the best ones to require or recommend... we deserve formal objections... Evidence that we avoided attempting the hard work is also useful for objections.

OR13:
I opened https://github.com/w3c/vc-wg-charter/pull/73, I suggest we close this PR, and if you approved it, move your approval there.

kdenhartog:
> > The tough part I see here is I don't actually think we'd be able to reach consensus on a "required" status crypto suite since we'd likely run back into the JWT vs JSON-LD debate and I could see that being used as evidence against the WG to object.
> 
> I think we should not run from objections we should run towards them, and address them.
> 
> I am sure that folks will try to make the case for a few required suites, and if we can't agree on the best ones to require or recommend... we deserve formal objections... Evidence that we avoided attempting the hard work is also useful for objections.

I'm alright with that reasoning, but it also misses the primary point of that post which is to say that this is better off defined as a CCG report that becomes a registry entry rather than a non-normative note because notes aren't protected by the patent policy and are "published to provide a stable reference for a useful document that is not intended to be a formal standard.". This is according to [section 6.4.1 of the W3C process document.](https://www.w3.org/2021/Process-20211102/#Note) As far as I know this is intended to be a formal standard, but we don't think we'll get consensus this go around so we should pull it out of scope. I'm raising it here since this seems like one of the items that I think is highly unlikely to go REC track in this WG this go around so I'm hoping that we can still keep it patent protected rather than try to give it some sort of symbolic status that seems to be misunderstood.

brentzundel:
I believe that if PR #77 is merged (in it's current state) this PR can be closed.

OR13:
I suggest closing the PR due to the recent changes in the section, its easier to add links in a new PR

msporny:
PR #77 has been merged, which includes PGP in the list of conditional normative specifications, closing.

PR #60, opened 2022-01-28
Clarified data model and sub parts

### LINKED ISSUE #57: Data model or models
I see a work item of a data model (singular) with words in the scope section about data models (plural).  What other data models are in scope?

brentzundel:
This PR tries to address #57 

***

### :boom: Error: 429 Too Many Requests :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on Feb 15, 2022, 3:30 PM UTC)_.

<details>
<summary>More</summary>

:link: [Related URL](https://rawcdn.githack.com/brentzundel/vc-wg-charter/3fdee4a7711f9f79abb0126b804b3ebd288a37f0/index.html)

```
429: Too Many Requests
```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/vc-wg-charter%2360.)._
</details>

### DIFF
--- a/index.html
+++ b/index.html
@@ -147,10 +147,15 @@ The <b>scope</b> of the Verifiable Credentials Working Group is:
         </p>
         <ul>
           <li>Addressing errata, ambiguities, and interoperability problems found in previous versions of the Verifiable Credentials Data Model</li>
-          <li>Data models for credentials, presentations, and proofs</li>
-          <li>Registries for the data models</li>
+          <li>A data model for Verifiable Credentials, inclusive of:</li>
+            <ul>
+                <li>credentials and verifiable credentials</li>
+                <li>presentations and verifiable presentations</li>
+                <li>requests for or submissions of verifiable credentials or verifiable presentations</li>
+            </ul>
+          <li>Registries for the data model</li>
           <li>Algorithms for the expression and verification of proofs that use existing cryptographic primitives</li>
-          <li>Refining multilingual support in data models</li>
+          <li>Refining multilingual support in the data model</li>
         </ul>
 	<p>
 	  It is explicitly not a requirement that the new specifications be fully compatible with related past specifications.


iherman (approved):

msporny (approved):
We lose the "proof" language in this PR. I think that's probably fine because of the large portion of input documents for cryptosuites, so it'll be very hard for someone to argue that we didn't contemplate the standardization of cryptosuites. Just noting the concern, I think the PR is fine as is.

brentzundel:
> We lose the "proof" language in this PR. I think that's probably fine because of the large portion of input documents for cryptosuites, so it'll be very hard for someone to argue that we didn't contemplate the standardization of cryptosuites. Just noting the concern, I think the PR is fine as is.

While `proofs` are no longer explicitly listed, I see them as a vital component of a data model for `credentials and verifiable credentials`, so I think we're covered.

brentzundel:
@Sakurann can you review?

index.html
```diff
@@ -147,10 +147,15 @@ <h2>Scope</h2>
...
           <li>Addressing errata, ambiguities, and interoperability problems found in previous versions of the Verifiable Credentials Data Model</li>
-          <li>Data models for credentials, presentations, and proofs</li>
-          <li>Registries for the data models</li>
+          <li>A data model for Verifiable Credentials, inclusive of:</li>
+            <ul>
+                <li>credentials and verifiable credentials</li>
+                <li>presentations and verifiable presentations</li>
+                <li>requests for or submissions of verifiable credentials or verifiable presentations</li>
+            </ul>
+          <li>Registries for the data model</li>
```
Sakurann, line 156: registry for which aspect of the data model? crypto suites?

index.html
```diff
@@ -147,10 +147,15 @@ <h2>Scope</h2>
         </p>
         <ul>
           <li>Addressing errata, ambiguities, and interoperability problems found in previous versions of the Verifiable Credentials Data Model</li>
-          <li>Data models for credentials, presentations, and proofs</li>
-          <li>Registries for the data models</li>
+          <li>A data model for Verifiable Credentials, inclusive of:</li>
+            <ul>
+                <li>credentials and verifiable credentials</li>
+                <li>presentations and verifiable presentations</li>
```
Sakurann, line 153: it's quite hard to know without the context why there is a differentiation between credentials and verifiable credentials; and between presentations and verifiable presentations. 
I think it would be helpful if we could clarify that we are defining a data model of a credentials, how it can be integrity protected/signed to be verifiable; and data model and integrity protection/signing of proof of possession (presentation and verifiable presentation)?

index.html
```diff
@@ -147,10 +147,15 @@ <h2>Scope</h2>
...
+                <li>requests for or submissions of verifiable credentials or verifiable presentations</li>
```
Sakurann, line 154: What is the intension of this line? it's already stated that we are defining a data model, will data model for "requests or submissions" be different from a "general" data model?

index.html
msporny: If the DID WG is any indication of the direction this will take, I expect the registry will be for all extension aspects of the VC Data Model. This is the current registry that we have:

https://w3c-ccg.github.io/vc-extension-registry/

We should probably list that as an input document to the VCWG.

index.html
msporny: > it's quite hard to know without the context why there is a differentiation between credentials and verifiable credentials; and between presentations and verifiable presentations.

The distinction between a credential and a verifiable credential is made here: https://w3c.github.io/vc-data-model/#credentials

The distinction between a presentation and a verifiable presentation is made here: https://w3c.github.io/vc-data-model/#presentations

> I think it would be helpful if we could clarify that we are defining a data model of a credentials

The current specification seems to do that already and the charter seems to indicate that we're going to continue doing that?

> how it can be integrity protected/signed to be verifiable

That is covered in the current specification here: https://w3c.github.io/vc-data-model/#proofs-signatures

... and the presumption is that we're going to refine that (thus the Data Integrity work item).

> data model and integrity protection/signing of proof of possession (presentation and verifiable presentation)?

I do think that's in scope as well.

The reason we list all of these things out above is because @David-Chadwick insisted that we be precise about what is included in the data model, which was a good idea.

Perhaps I'm missing your point @Sakurann... could you provide concrete text that we might use in addition to/in place of the text above?

index.html
msporny: As stated above (https://github.com/w3c/vc-wg-charter/pull/60#discussion_r799477016), we are trying to be more precise about what is in scope for the VCWG. Specifically, the line above covers "presentation requests" -- that is, "how does one request a verifiable presentation/credential?" The VCWG 1.0 covered how you express a presentation and verifiable presentation, but it did not cover how a verifier requests one (though, some those mechanisms  ([VPR](https://w3c-ccg.github.io/vp-request-spec/), [WACI/PeX](https://identity.foundation/waci-presentation-exchange/), [Aries Present Proof](https://github.com/hyperledger/aries-rfcs/tree/main/features/0454-present-proof-v2)) have existed for 6 years, we were waiting for the market to mature to the point where we could have multiple inputs in order to standardize them).

index.html
OR13: Related, but not the same this: https://github.com/w3c-ccg/ld-cryptosuite-registry

Sakurann (approved):
Approve to restructure in a proposed way.
related to the comments/discussions on this PR, I do think we need PRs to 
- point to certain definitions in vc-data-model-v.1.1
- clearer language about the registries

PR #61, opened 2022-01-28
Proposed work mode for charter development

brentzundel:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,7 +33,7 @@ required, particularly while this document is in a draft state.
 PRs will be merged at the chairs' discretion, and may be merged quickly. Please
 raise an issue if changes have been made to the draft charter that you do not
 agree with. However, please keep in mind that while we will strive to address
-issues in a manner that is acceptable to those have raised them, we cannot
+issues in a manner that is acceptable to those who have raised them, we cannot
 guarantee that all issues will result in changes satisfactory to all parties,
 nor that all parties will have sufficient time to review changes before they are
 merged into the draft.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,6 @@
 # Verifiable Claims Working Group
 
+## Contributions
 Contributions to this repository are intended to become part of Recommendation-track documents governed by the
 [W3C Patent Policy](https://www.w3.org/Consortium/Patent-Policy-20040205/) and
 [Document License](https://www.w3.org/Consortium/Legal/copyright-documents). To make substantive contributions to specifications, you must either participate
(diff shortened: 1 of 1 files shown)


iherman (approved):

CONTRIBUTING.md
```diff
@@ -22,3 +23,18 @@ If you added a contributor by mistake, you can remove them in a comment with:
...
+the listed chairs before merging. Other approvals may be sought, but are not
+required, particularly while this document is in a draft state.
+
+PRs will be merged at the chairs' discretion, and may be merged quickly. Please
+raise an issue if changes have been made to the draft charter that you do not
+agree with. However, please keep in mind that while we will strive to address
+issues in a manner that is acceptable to those have raised them, we cannot
+guarantee that all issues will result in changes satisfactory to all parties,
+nor that all parties will have sufficient time to review changes before they are
+merged into the draft.
```
msporny: > PRs will be merged at the chairs' discretion, and may be merged quickly. 

The VCWG Charter is a VCWG work item, isn't it? Why are we not allowing VCWG members enough time to review PRs? This creates an atmosphere where things can be rushed through, which could then just inevitably lead to objections when the charter goes out to review.

My suggestion is that PRs that are not Editorial in nature must be given time on at least one call for discussion, and a minimum of 7 days from the day they were raised, to be merged by the Chairs.

Is there some kind of deadline that we're working against that is driving this sort of work mode?

msporny (changes requested):
The VCWG Charter is a VCWG work item, isn't it? Why are we not ensuring that VCWG members have enough time to review PRs to a charter to a WG that they will be participating in?

This creates an atmosphere where things can be rushed through, which could then just inevitably lead to objections when the charter goes out to review. I don't understand the rush here, what did I miss?

My suggestion is that PRs must be given time on at least one call for discussion, and a minimum of 7 days from the day they were raised, to be merged by the Chairs.

CONTRIBUTING.md
brentzundel: I'd like to see the charter draft ready for AC review within a couple of weeks, which imo doesn't allow for review periods, 
I have added a line which I hope allays concerns that the chairs will submit a charter which does not meet with the approval to the AC for review.

msporny (approved):
> Before the draft charter is sent for Advisory Committee review it will be ￼submitted to the current VCWG for approval.

I'm now ok w/ this work mode as long as the VCWG reaches consensus on the final charter before submission to the AC for voting.

brentzundel:
@Sakurann I am hoping to get your opinion on this proposed work mode.

kdenhartog (approved):

Sakurann (approved):
I think this Work Mode makes sense. Thank you, Brent.

Issue #64, opened 2022-01-31
v2 Test Suite


OR13:
I would like to define our objectives for the v2 test suite more clearly, before asking for approval on the charter to "work on test suites".

Specifically I would like to see test suites implemented that address the following:

1. issuance and verification with private key and public key material disclosed in test vectors.
2. data model conformance tests for "production and consumption" between JSON and JWT, LD Proof, JWP, etc.
3. demonstrate agility for proof mechanisms (ability to change cryptographic primitives without changing the data model).

OR13:
We've been working on addressing some of these issues here: https://github.com/decentralized-identity/JWS-Test-Suite

Specifically VC-JWT and JsonWebSignature2020 LD Proof interoperability.

selfissued:
Improving the precision of the deliverable in this way makes sense to me.

OR13:
We should include a link under the other deliverables section that explains how to test VC-JWT and JsonWebSignature2020 side by side.

brentzundel:
I believe this issue has been addressed. Marking as `pending close`.
I will close this issue in 4 days if there are no objections.

brentzundel:
no response since marked `pending close`, closing

Issue #54, opened 2022-01-27
Update the Verifiable Credential Data Integrity 1.0 section


OR13:
The entire section needs to be rewritten to communicate layers in a less ambiguous manner and cite references respectfully.

As the sole author of the following:

- [Ecdsa Secp256k1 Signature 2019](https://w3c-ccg.github.io/lds-ecdsa-secp256k1-2019/)
- [Ed25519 Signature 2020](https://w3c-ccg.github.io/lds-ed25519-2020/)

Co-editor / author of:

- [BBS+ Signatures 2020](https://w3c-ccg.github.io/ldp-bbs2020/)

Contributor to:

- [draft-jmiller-json-proof-token](https://json-web-proofs.github.io/json-web-proofs/)

I find the "rebranding" of the work items completely unacceptable.

The work items MUST be referred to by the names that they are currently using, anything else is unacceptable.

Additionally, there is a harmful ambiguity that is being perpetuated by this section which is conflating "Data Integrity" with "cryptographic serializations" with "raw cryptographic primitives".

I think its fine to refer to each layer in this section, but 100% not acceptable to "group everything together as if these are the same things".

Our charter forbids us from doing "raw cryptographic primitives", yet we can refer to work happening elsewhere.

Our charter grants us the ability to do "cryptographic serializations", but what we do here and what we happens elsewhere MUST be clearly and cleanly separated in unambiguous language.

Our charter grants use the ability to do "Data Integrity", but not to define "one ring to rule them all" form of it... We need to clearly communicate how this section will account for supporting decentralization and diversity.

selfissued:
> I find the "rebranding" of the work items completely unacceptable.

I completely agree with this.  References to all referenced specifications must use the accurate titles for these specifications.

bumblefudge:
Uh oh, apologies if my PR contributed to this.  Go easy on the new guy!

brentzundel:
@OR13 am I correct in assuming that this issue may be closed if PR #77 is merged?

brentzundel:
I believe this issue has been addressed. Marking as `pending close`.
I will close this issue in 4 days if there are no objections.

brentzundel:
no response since marked `pending close`, closing

PR #73, opened 2022-02-15
Add links from stalled prs

OR13:
- move pgp to other deliverables section per feedback on https://github.com/w3c/vc-wg-charter/pull/50
- move multikey to other deliverables section per feedback on https://github.com/w3c/vc-wg-charter/pull/51
- add json web signature test suite per https://github.com/w3c/vc-wg-charter/issues/64

***
<a href="https://pr-preview.s3.amazonaws.com/OR13/vc-wg-charter/pull/73.html" title="Last updated on Feb 15, 2022, 8:04 PM UTC (9e2a1b1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-wg-charter/73/9a0ef6b...OR13:9e2a1b1.html" title="Last updated on Feb 15, 2022, 8:04 PM UTC (9e2a1b1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -275,10 +275,17 @@ produced.
                 <li>Test Suites</li>
               </ul>
             </li>
+            <li>Other VC Data Integrity proof types:
+              <ul>
+                <li>VC Data Integrity - <a href="https://or13.github.io/lds-pgp2021"></a>PGP</li>
+                <li>VC Data Integrity - <a href="https://ns.did.ai/suites/multikey-2021"></a>Multbase / MultiCodec</li>
+              </ul>
+            </li>
             <li>Test suites for verifiable credential proof types:
             <ul>
               <li>VC Data Integrity - Ed25519</li>
               <li>VC Data Integrity - BBS+</li>
+              <li>VC Data Integrity - Json Web Signatures</li>
               <li>VC JSON Web Tokens</li>
             </ul>
             </li>
--- a/index.html
+++ b/index.html
@@ -278,7 +278,6 @@ produced.
             <li>Other VC Data Integrity proof types:
               <ul>
                 <li>VC Data Integrity - <a href="https://or13.github.io/lds-pgp2021"></a>PGP</li>
-                <li>VC Data Integrity - <a href="https://ns.did.ai/suites/multikey-2021"></a>Multbase / MultiCodec</li>
               </ul>
             </li>
             <li>Test suites for verifiable credential proof types:


index.html
```diff
@@ -275,10 +275,17 @@ <h3>
                 <li>Test Suites</li>
               </ul>
             </li>
+            <li>Other VC Data Integrity proof types:
+              <ul>
+                <li>VC Data Integrity - <a href="https://or13.github.io/lds-pgp2021"></a>PGP</li>
+                <li>VC Data Integrity - <a href="https://ns.did.ai/suites/multikey-2021"></a>Multbase / MultiCodec</li>
```
msporny: This is the only line I would object to. We put `publicKeyMultibase` and Multikey in DID Core with the explicit goal to finish standardizing it in the VC 2.0 WG. This is a normative deliverable, to be defined in Data Integrity, and must remain as such.

msporny (changes requested):
I only object to one line in the PR (that sends a mixed message). Remove it and I will approve.

index.html
OR13: >  This is a normative deliverable, to be defined in Data Integrity, and must remain as such.

I didn't touch that part of the charter, which already covers Ed25519 2020 and with your PR (https://github.com/w3c/vc-wg-charter/pull/63) will add Secp384r1 2022.

Until the normatively deliverables cover other popular key formats (like P-256 / ES256) which is mandatory here: https://www.iana.org/assignments/jose/jose.xhtml

I think its smart for us to keep the door open to working on them non normatively.

OR13:
@msporny either update your PR (https://github.com/w3c/vc-wg-charter/pull/63) to cover P-256 (and others), or accept this PR to allow us to define them non normatively... we can't skip over mandatory to implement key and signature formats and expect to have the charter approved.

Another alternative would be to move the LD specific key and signature representations that are not based on JWS to the section in this PR.

I don't consider the current attempts at defining and omitting specific keys and signatures in this WG to be a path to success.

It's going to lead to formal objections for not addressing the topic consistently. 

Either `publicKeyMultibase` is a viable alternative to `publicKeyJwk`, or our charter should close the door on it, and allow us all to focus on other components of this specification.

msporny:
> Either `publicKeyMultibase` is a viable alternative to `publicKeyJwk`, or our charter should close the door on it, and allow us all to focus on other components of this specification.

Can you please explain how what is proposed in this spec (normative deliverable of WG) doesn't make `publicKeyMultibase` a viable alternative to `publicKeyJwk`?

* https://w3c-ccg.github.io/data-integrity-spec/#multikey
* https://w3c-ccg.github.io/data-integrity-spec/#verification-material (see ISSUE 8)

The goal there is to define all possible key types in the multicodec table, which include (IIRC) all key types that are possible to be expressed in `publicKeyJwk`. You asked me to write spec text to enable that, I did that, and now you are still suggesting that that didn't happen. Have you read the spec text? What am I missing here? 

Please state, concretely, why you believe the Multikey that you want is not supported via a normative work item that is currently listed in the charter?

OR13:
@msporny your PR #63 

is titled: 

> [DO NOT MERGE YET] Add ECDSA secpr1 cryptosuite to VCWG input documents.

And it points to:

> https://digitalbazaar.github.io/di-ecdsa-secpr1-2019/

Which in turn looks like this:

Screen Shot 2022-02-15 at 12 05 59 PM

https://digitalbazaar.github.io/di-ecdsa-secpr1-2019/#ecdsasecp256r1signature2019

I suppose we might consider this "close enough"... but its not defining:

- Ed25519
- Secp256k1
- Bls12381
- RSA

I guess if we add up all the CCG specs for key types and signatures, we do indeed have a working solution (except for RSA and P-521).

My point was that its pretty painful to have to review:

- https://digitalbazaar.github.io/di-ecdsa-secpr1-2019/ ( Secp256r1 + Secp384r1 (and not P-521) )
- https://w3c-ccg.github.io/ldp-bbs2020/ (Bls12381)
- https://w3c-ccg.github.io/lds-ed25519-2020/ (Ed25519)
- https://w3c-ccg.github.io/lds-ecdsa-secp256k1-2019/ (Secp256k1)
- no spec for RSA

If you are saying that this is just a function of where the input documents are today / how they were created, and that indeed our intention truly is to solve this `key type` and `signature type` for all entries in:

https://github.com/multiformats/multicodec/blob/master/table.csv#L90

We are aligned, but this feels significantly less concrete than a single document stating this clearly.

We have such a document, here: https://w3c-ccg.github.io/data-integrity-spec/#multikey

Perhaps we don't need all these suites in the "normatively defined" section if we have the link above? (which is already in the normative section?)

This PR (#73) makes it clear that we do intend to address these concerns at least non normatively for existing crypto that we are using today.

I will be happy to remove multikey from the non normative section, or approve your PR if it removes it from this section in the future.

If you require me to remove a non normative reference to the key format we use to address these issues currently, which is  already conformant with `https://w3c-ccg.github.io/data-integrity-spec/#multikey` modulo choices for `names`, I can do that.

index.html
OR13:
```suggestion
```

OR13: @msporny please confirm this is the change you require.

msporny: Yes, confirmed... and to be crystal clear so comment thread is not used against the VCWG: The reason I'm recommending the removal of this line item is because multikey, which uses multibase and multicodec, is already a part of a normative deliverable the group has identified in the current charter proposal. To say that it is both a normative and non-normative deliverable would be confusing. We should state clearly that it is a normative deliverable (as a part of the verification method definition in the Data Integrity specification).

I'll respond to your commentary above when I get some free cycles, but the TL;DR is that I believe we're aligned on the end goal here (or some variation of it).

OR13:
@msporny I have accepted your change suggestion, we can tackle the rest of the commentary on this thread via this issue:

https://github.com/w3c/vc-wg-charter/issues/54

msporny (approved):

selfissued (commented):
Looking at the diffs, it doesn't appear that this adds a reference to the test suite.  I'd prefer that we did.

brentzundel:
@OR13 what is the status of this PR? Does it still reflect changes you feel are needed, or would it be best to see what happens with PR #77 before merging?

Also, I share Mike's question about a reference to the test suite and agree that is should be included.

index.html
```diff
@@ -275,10 +275,16 @@ <h3>
...
+            <li>Other VC Data Integrity proof types:
+              <ul>
+                <li>VC Data Integrity - <a href="https://or13.github.io/lds-pgp2021"></a>PGP</li>
+              </ul>
+            </li>
             <li>Test suites for verifiable credential proof types:
             <ul>
               <li>VC Data Integrity - Ed25519</li>
               <li>VC Data Integrity - BBS+</li>
+              <li>VC Data Integrity - Json Web Signatures</li>
```
msporny, line 287: We should collapse all items above to "Test Suites for WG-defined crypographic container formats and suites."

OR13:
closing due to instability in the section, I will take another stab when it settles.

Issue #67, opened 2022-02-03
Each registry should include at least one standardized entry


jyasskin:
I see that the charter includes "Registries for the data models", but that might not be interpreted as including a definition of the entries in those registries. For example, https://w3c-ccg.github.io/vc-extension-registry/ includes 4 empty registries, which wouldn't be sufficient to make the data model a good specification.

kdenhartog:
I agree with this. Historically speaking we were still trying to get a better understanding of how registries would work in W3C around the time that this extension registry was formed. As a part of that this registry hasn't operated quite like how the DID Core registry has but I'd expect going forward that it would be a lot more useful or be removed. With that in mind, do you have particular additional text you want to see added here to make sure this doesn't happen again?

jyasskin:
Maybe

> Scope:
> * ...
> * Registries for the data models
> * Enough REC-track specifications to ensure each registry has enough standards-track entries to exercise the expected range of variation for that registry.

Or it could be less precise:

> Scope:
> * ...
> * Registries for the data models and REC-track specifications to appear in those registries.

kdenhartog:
Ahh I see what you're bringing up now. We're defining extension points without even having one recommendation-track default. This could expand scope considerably if we make it recommendation-track for every point (leaves the question of does level 2 need to drop features from v1 that don't have a rec-track approach?), but I think it's a wise point to bring up even though I suspect it's going to raise some controversy.

iherman:
I understand the direction your going @jyasskin but I am a bit worried. Indeed, if we consider 'recommendation-track' default then the immediate question is: which working group would have this in its charter? And currently the answer might be: none. Ie, this can work only if we extend _this_ charter by adding those recommendations, and I do not think that would be accepted.

Also (although I admit this is a minor point), the term 'recommendation' is W3C specific, and we may get such documents from other organizations.

I am not absolutely sure what to replace this with. We may say 'stable and open source specifications', or 'stable specifications that have several mutually independent implementations', or something like that (although it is quite verbose).

jyasskin:
I want to avoid a situation like we had with DID-core where we couldn't review the core spec's full behavior because no standards-track WG had endorsed any of the implementations of a critical extension point. Having only non-standards-track implementations of some extension points also makes it hard for the core REC to effectively foster interoperability, since implementations will have to guess at how to implement those extension points.

[...]

PR #98, opened 2022-03-08
The WG may produce registries

brentzundel:
This PR takes one part of PR #85 which I hope will achieve consensus.

***
<a href="https://pr-preview.s3.amazonaws.com/brentzundel/vc-wg-charter/pull/98.html" title="Last updated on Mar 8, 2022, 6:51 PM UTC (5115992)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-wg-charter/98/bb419e2...brentzundel:5115992.html" title="Last updated on Mar 8, 2022, 6:51 PM UTC (5115992)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -306,8 +306,11 @@ disclosure and other modern cryptographic schemes.
             Registries
           </h3>
           <p>
-A series of at least the following registries that support the normative
-deliverables of the Working Group will be published and maintained:
+The Working Group will deliver a set of registries on the [W3C Registry
+Track](https://www.w3.org/2021/Process-20211102/#registries) to
+support extension points in the above normative deliverables. The
+Working Group expects to produce the following registries, but it will
+adjust this list as needed to support the normative deliverables:
           </p>
           <table>
             <tr>
--- a/index.html
+++ b/index.html
@@ -306,8 +306,7 @@ disclosure and other modern cryptographic schemes.
             Registries
           </h3>
           <p>
-The Working Group will deliver a set of registries on the [W3C Registry
-Track](https://www.w3.org/2021/Process-20211102/#registries) to
+The Working Group will deliver a set of registries on the <a href="https://www.w3.org/2021/Process-20211102/#registries">W3C Registry Track</a> to
 support extension points in the above normative deliverables. The
 Working Group expects to produce the following registries, but it will
 adjust this list as needed to support the normative deliverables:
--- a/index.html
+++ b/index.html
@@ -306,10 +306,11 @@ disclosure and other modern cryptographic schemes.
             Registries
           </h3>
           <p>
-The Working Group will deliver a set of registries on the <a href="https://www.w3.org/2021/Process-20211102/#registries">W3C Registry Track</a> to
-support extension points in the above normative deliverables. The
-Working Group expects to produce the following registries, but it will
-adjust this list as needed to support the normative deliverables:
+The Working Group may deliver a set of registries on the
+<a href="https://www.w3.org/2021/Process-20211102/#registries">W3C Registry Track</a>
+to support extension points in the above normative deliverables. The Working
+Group might produce the following registries and will adjust this list
+as needed to support the normative deliverables:
           </p>
           <table>
             <tr>
--- a/index.html
+++ b/index.html
@@ -320,10 +320,11 @@ adjust this list as needed to support the normative deliverables:
             </tr>
             <tr>
               <td>
-Verifiable Credential Extensions Registry
+Verifiable Credential Properties Registry
               </td>
               <td>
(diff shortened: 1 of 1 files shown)


msporny (approved):
I would personally like to be more specific, but would be fine w/ this text as well.

brentzundel:
I've also added text which addresses Issue #67

index.html
```diff
@@ -301,6 +301,21 @@ <h3>
...
+            Registries
+          </h3>
+          <p>
+The Working Group may create a set of
+<a href="https://www.w3.org/2021/Process-20211102/#registries">registries</a>
+including <a href="https://www.w3.org/2021/Process-20211102/#registry-definition">registry definitions</a>
+and <a href="https://www.w3.org/2021/Process-20211102/#registry-table">registry tables</a>
+to support extension points in the above normative deliverables. Registries for
+extension points that are required by any of the above normative deliverables
+must have at least one standardized entry.
```
msporny:
```suggestion
to support extension points in the above normative deliverables. Registries for
extension points that are required by any of the above normative deliverables
must have at least one standardized entry.
```

-1, I expect that we'd have to remove features in the VC spec and that would be controversial.

msporny (changes requested):
I was supportive of this PR, but now am not given the recent change in 6753b9e1396e34ee43f22ec1cecdb4cca7699eb4 and confusion in https://github.com/w3c/vc-wg-charter/pull/98#discussion_r821959266.

index.html
msporny: Hrm, on second read, I'm confused. So, existing extension points that don't have normative registry entries would still be allowed?

brentzundel: I tried to add the language that seemed to have the most consensus during the WG call on March 2nd.
Currently, the only normatively required extension point in the VC data model is 'proof', which we will standardize in the data integrity work. I don't see any features that we would need to remove. What am I missing?

brentzundel:
@msporny since the original intent of this PR was to try and find consensus on the WG having any registries, I have removed the controversial change and will reintroduce it in another PR

index.html
msporny: > I don't see any features that we would need to remove. What am I missing?

We currently don't have normative deliverables in the current charter defined for any of the following extension points:

https://www.w3.org/TR/vc-data-model/#status
https://www.w3.org/TR/vc-data-model/#evidence
https://www.w3.org/TR/vc-data-model/#terms-of-use
https://www.w3.org/TR/vc-data-model/#refreshing
https://www.w3.org/TR/vc-data-model/#data-schemas

One reading of the language wrt. "normative registry entries" is that every single one of those features should be removed in the VC 2.0 work, which I expect would lead to objections from organizations relying on those extension points out in the field.

brentzundel: none of those are required by the VC data model, therefore none of those would need a standardized entry in a registry.
But also, the language has been removed form this PR and moved to #101

msporny: > none of those are required by the VC data model, therefore none of those would need a standardized entry in a registry.

Ah, I see, and am now slowly remembering the conversation... I'm not opposed IF everyone reads the language in the same way. I'll try providing some changes in PR #101 to try to make it clear that it only applies to normative properties... though, I expect, this wouldn't address @jyasskin's concern.

selfissued (approved):
This is great!  This is all we need to say about registries in the charter.

kdenhartog (approved):

Sakurann (approved):

iherman (approved):

OR13 (approved):

msporny (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-03-09-vcwg#section1-2) on 2022-03-09

- no resolutions were taken

#### 1.2. The WG may produce registries (pr vc-wg-charter#98)

_See github pull request [vc-wg-charter#98](https://github.com/w3c/vc-wg-charter/pull/98)._




**Brent Zundel:** looks like we have rough agreement, we want registries in the charter for the most part.  
… can you throw your approval instead of the stale CR, any opposition to merging this PR.  

> *Joe Andrieu:* I oppose, for reasons stated.

**Brent Zundel:** Joe your opposition is noted, and will go into the comments on the PR.  

**Oliver Terbu:** does registries include data integrity proof registry, my concern is it should be permissive enough, and if we can get that right then I'm on board.  

> *Manu Sporny:* +1 to registry not being restrictive... it needs to be permissive..

**Brent Zundel:** the charter doesn't restrict as of yet but the vcwg may restrict at a later date.  
… over the noted objection from joe will merge the PR.  
… two other PRs that build towards what 85 was aiming for, looking at 99 next..  

</details>

PR #101, opened 2022-03-08
Registries and standardized entries

brentzundel:
This PR build on PR #98 to address Issue #67. 

It adds language that requires registry entries for normatively required extensions to have at least one standardized entry.

***
<a href="https://pr-preview.s3.amazonaws.com/brentzundel/vc-wg-charter/pull/101.html" title="Last updated on Mar 8, 2022, 8:44 PM UTC (99e4ca1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-wg-charter/101/bb419e2...brentzundel:99e4ca1.html" title="Last updated on Mar 8, 2022, 8:44 PM UTC (99e4ca1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -310,7 +310,9 @@ The Working Group may create a set of
 <a href="https://www.w3.org/2021/Process-20211102/#registries">registries</a>
 including <a href="https://www.w3.org/2021/Process-20211102/#registry-definition">registry definitions</a>
 and <a href="https://www.w3.org/2021/Process-20211102/#registry-table">registry tables</a>
-to support extension points in the above normative deliverables.
+to support extension points in the above normative deliverables. Registries for
+extension points that are required by any of the above normative deliverables
+must have at least one standardized entry.
           </p>
         </section>
 
--- a/index.html
+++ b/index.html
@@ -306,8 +306,11 @@ disclosure and other modern cryptographic schemes.
             Registries
           </h3>
           <p>
-A series of at least the following registries that support the normative
-deliverables of the Working Group will be published and maintained:
+The Working Group will deliver a set of registries on the [W3C Registry
+Track](https://www.w3.org/2021/Process-20211102/#registries) to
+support extension points in the above normative deliverables. The
+Working Group expects to produce the following registries, but it will
+adjust this list as needed to support the normative deliverables:
           </p>
           <table>
             <tr>
--- a/index.html
+++ b/index.html
@@ -306,8 +306,7 @@ disclosure and other modern cryptographic schemes.
             Registries
           </h3>
           <p>
-The Working Group will deliver a set of registries on the [W3C Registry
-Track](https://www.w3.org/2021/Process-20211102/#registries) to
+The Working Group will deliver a set of registries on the <a href="https://www.w3.org/2021/Process-20211102/#registries">W3C Registry Track</a> to
 support extension points in the above normative deliverables. The
 Working Group expects to produce the following registries, but it will
 adjust this list as needed to support the normative deliverables:
--- a/index.html
+++ b/index.html
@@ -306,10 +306,11 @@ disclosure and other modern cryptographic schemes.
             Registries
           </h3>
           <p>
-The Working Group will deliver a set of registries on the <a href="https://www.w3.org/2021/Process-20211102/#registries">W3C Registry Track</a> to
-support extension points in the above normative deliverables. The
-Working Group expects to produce the following registries, but it will
-adjust this list as needed to support the normative deliverables:
+The Working Group may deliver a set of registries on the
+<a href="https://www.w3.org/2021/Process-20211102/#registries">W3C Registry Track</a>
+to support extension points in the above normative deliverables. The Working
+Group might produce the following registries and will adjust this list
+as needed to support the normative deliverables:
           </p>
           <table>
             <tr>
--- a/index.html
+++ b/index.html
@@ -320,10 +320,11 @@ adjust this list as needed to support the normative deliverables:
             </tr>
             <tr>
               <td>
-Verifiable Credential Extensions Registry
+Verifiable Credential Properties Registry
               </td>
               <td>
-A registry of extensions to the Verifiable Credentials Data Model specification.
+A registry of extensions to properties used in the Verifiable Credentials Data
+Model specification.
               </td>
               <td>
 <a href="https://w3c-ccg.github.io/vc-extension-registry/">Verifiable Credential Extensions Registry</a>
@@ -335,7 +336,7 @@ Cryptographic Container and Suite Registry
               </td>
               <td>
 A registry of cryptographic container formats and cryptographic suites that
-are useful in conjunction with specifications managed by this Working Group
+are useful in conjunction with specifications managed by this Working Group.
               </td>
               <td>
 <a href="https://w3c-ccg.github.io/security-vocab/#classes">Classes in CCG Security Vocabulary</a>
--- a/index.html
+++ b/index.html
@@ -339,7 +339,7 @@ A registry of cryptographic container formats and cryptographic suites that
 are useful in conjunction with specifications managed by this Working Group.
               </td>
               <td>
-<a href="https://w3c-ccg.github.io/security-vocab/#classes">Classes in CCG Security Vocabulary</a>
+<a href="https://w3c-ccg.github.io/security-vocab/#classes">Classes in Security Vocabulary</a>
               </td>
             </tr>
             <tr>
--- a/index.html
+++ b/index.html
@@ -316,38 +316,51 @@ adjust this list as needed to support the normative deliverables:
             <tr>
               <th>Registry</th>
               <th>Description</th>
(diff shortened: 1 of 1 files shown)


msporny (changes requested):
We currently don't have normative deliverables in the current charter defined for any of the following extension points:

https://www.w3.org/TR/vc-data-model/#status
https://www.w3.org/TR/vc-data-model/#evidence
https://www.w3.org/TR/vc-data-model/#terms-of-use
https://www.w3.org/TR/vc-data-model/#refreshing
https://www.w3.org/TR/vc-data-model/#data-schemas

One reading of the language in this PR wrt. "normative registry entries" is that every single one of those features should be removed in the VC 2.0 work, which I expect would lead to objections from organizations relying on those extension points out in the field. I have no alternate language or action to propose, other than closing this PR.

brentzundel:
The language introduced in this PR would require standardized entries for normatively required extension points.

None of the extension points you mentioned are normatively required, so none of them would need to have a standardized entry, i.e., none of them would need to be removed from the spec.

msporny:
> None of the extension points you mentioned are normatively required, so none of them would need to have a standardized entry, i.e., none of them would need to be removed from the spec.

Ah, now I see the intent of the language. I'm not opposed to the premise, though I don't think it would address @jyasskin's concerns? He will have to weigh in on that.

I'll try to propose some alternate language that makes it clear what "normatively required" means, but before I do that... `proof` is only normatively required for embedded proofs, so even for that, there's an argument that we don't have to define anything normative (even though we're going to do that).

index.html
```diff
@@ -301,6 +301,21 @@ <h3>
...
+            Registries
+          </h3>
+          <p>
+The Working Group may create a set of
+<a href="https://www.w3.org/2021/Process-20211102/#registries">registries</a>
+including <a href="https://www.w3.org/2021/Process-20211102/#registry-definition">registry definitions</a>
+and <a href="https://www.w3.org/2021/Process-20211102/#registry-table">registry tables</a>
+to support extension points in the above normative deliverables. Registries for
+extension points that are required by any of the above normative deliverables
+must have at least one standardized entry.
```
msporny: Alternate # 1 (more terse): 
```suggestion
to support extension points in the above normative deliverables. Registries for
extension points that are mandatory to use, for any of the above normative 
deliverables, must have at least one standardized entry.
```

Alternate # 2 (wordier): 
```suggestion
to support extension points in the above normative deliverables. Registries for
extension points that are mandatory to use, for any of the above normative 
deliverables (for example, Verifiable Credential properties that MUST be included in 
a Verifiable Credential), must have at least one standardized entry.
```

msporny (changes requested):
Ok with this given the proposed changes are accepted.

OR13 (changes requested):
same comment as manu's

jyasskin:
This interacts with https://github.com/w3c/w3process/issues/597, where @dwsinger suggested that "we should try to write the (normative) registry definitions for those registries, and see what we come up with." I'm happy with that approach: accept @msporny's position for the charter, and then see if the resulting specification looks reasonable. That does include the risk that it'll wind up looking unreasonable in the Proposed REC and result in a formal objection then, so please remind me and the Process CG folks to take a look before that.

selfissued (approved):
I support this PR.  I believe that it achieves @jyasskin 's goals in https://github.com/w3c/vc-wg-charter/issues/67.

index.html
```diff
@@ -301,6 +301,22 @@ <h3>
...
+          </h3>
+          <p>
+The Working Group may create a set of
+<a href="https://www.w3.org/2021/Process-20211102/#registries">registries</a>
+including <a href="https://www.w3.org/2021/Process-20211102/#registry-definition">registry definitions</a>
+and <a href="https://www.w3.org/2021/Process-20211102/#registry-table">registry tables</a>
+to support extension points in the above normative deliverables. Registries for
+extension points that are mandatory to use, for any of the above normative 
+deliverables (for example, Verifiable Credential properties that MUST be included in 
+a Verifiable Credential), must have at least one standardized entry.
```
kdenhartog, line 316:
```suggestion
a Verifiable Credential), MUST have at least one standardized entry.
```

Not sure if this requires normative text to define so just wanted to point it out as an option.

Sakurann (approved):

index.html
```diff
@@ -301,6 +301,22 @@ <h3>
...
+        <section id="normative">
+          <h3>
+            Registries
+          </h3>
+          <p>
+The Working Group may create a set of
+<a href="https://www.w3.org/2021/Process-20211102/#registries">registries</a>
+including <a href="https://www.w3.org/2021/Process-20211102/#registry-definition">registry definitions</a>
+and <a href="https://www.w3.org/2021/Process-20211102/#registry-table">registry tables</a>
+to support extension points in the above normative deliverables. Registries for
```
iherman, line 313: (Purely editorial) It may be that my non-native English is the problem, but the formulation sounds a bit odd to my ears... 
```suggestion
to support extension points in the normative deliverables listed in the previous sections. Registries for
```

index.html
```diff
@@ -301,6 +301,22 @@ <h3>
...
+extension points that are mandatory to use, for any of the above normative 
+deliverables (for example, Verifiable Credential properties that MUST be included in
```
iherman, line 315: Same as the previous comment
```suggestion
extension points that are mandatory to use for any of the normative 
deliverables (for example, Verifiable Credential properties that MUST be included in 
```

iherman (approved):

index.html
msporny: > Not sure if this requires normative text to define so just wanted to point it out as an option.

W3C Charters typically do not use normative spec language. Doing so might raise a few eyebrows within the W3C AC during the vote. I suggest we stay away from [BCP14](https://www.rfc-editor.org/info/bcp14) language.

kdenhartog: Thanks, will note that info in the back of my head as well.

kdenhartog (approved):

OR13 (approved):

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2022-03-09-vcwg#section1-4) on 2022-03-09

- no resolutions were taken

#### 1.4. Registries and standardized entries (pr vc-wg-charter#101)

_See github pull request [vc-wg-charter#101](https://github.com/w3c/vc-wg-charter/pull/101)._




**Brent Zundel:** will spend the last few minutes on 101 instead of 100.  
… this adds a line to the section we just merged, that says " registries for extension points that are mandatory to use, for any of the above normative deliverables (for example, Verifiable Credential properties that MUST be included in ... must have one standardized entry".  

**Kyle Den Hartog:** a must for required properties and a should for optional, allows us to get to a more solid ground. This lets us define how extension points work in an interop way.  

**Joe Andrieu:** My question is what does standardized mean?.  

> *Kyle Den Hartog:* +1.

**Joe Andrieu:** all but one SO calls their standards standard, we need to clarify that language to know what it means..  

> *Kyle Den Hartog:* we can replace with REC Track document.

> *Kyle Den Hartog:* ahh yeah makes sense.

> *Dave Longley:* "REC track or equivalent level?".

**Brent Zundel:** we cant say rec track because it might be from IETF or ISO, there is a general understanding and it may be defined somewhere.  

> *Ivan Herman:* See ["normative references" for W3C](https://www.w3.org/2013/09/normative-references).

**david:** don't think it makes sense to have a must or a should for registries, if you take terms of use the semantics of it is very broad..  
… to say you should/must doesn't make sense when the semantics are wide..  

> *Orie Steele:* Should we even be using BCP14 language on ANY charter item? - [https://github.com/w3c/vc-wg-charter/pull/101/files#r822616764](https://github.com/w3c/vc-wg-charter/pull/101/files#r822616764).

**Manu Sporny:** we can use the same definition, for making a normative reference in the w3c and thats a well known tried and true mechanism. I agree with david chadwick said, I was under the same impression that terms of use status etc, but this PR does not apply to them. This only applies to the proof block. We're saying if you're using an embedded proof, you need ot use a REC track entry. IDK if others had this same understanding, this applies to almost no entry except the proof block.  

**Ivan Herman:** manu almost said but [normative reference in w3c recommendations](https://www.w3.org/2013/09/normative-references) says which kind of document I can refer to normatively, which covers the various things Joe wants, and that is very appropriate add a reference to in the charter..  

**Kyle Den Hartog:** for this PR its probably beneficial to not talk about the should but leave open the issue for optional properties, i think there's consensus on the must aspect. I think the issuer property needs to be more strongly defined, in how we dereference it to Public key materials, which will affect the proof properties.  

**Joe Andrieu:** this does only apply to normatively defined properties but it binds the group in what i think is a bad idea. it means we can't define a normative field that doesn't have a predefined process. I think making it mandatory is premature..  

</details>

msporny (approved):

brentzundel:
Changes requested and made, approved by both listed chairs, merging according to our current work mode.

